PR #8, opened 2017-04-07
Cache implementations

fehlmach:
created cache interfaces, refactoring switchboard, inbox and verify with new caches.
impl of cache interfaces: keyCache, ConnectionCache, CapabilityCache, pendingSignatureCache

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/8)


fehlmach:
PendingQueryCache, AssertionCache and negativeAssertionCache interfaces are not yet ready for review

---

Review status: 0 of 13 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-:-Kh67kQatCWvR-jNVw7u:bqbu6o1)*

klausman:
Reviewed 13 of 13 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[rainsd/engine.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6aYAWwriA9Sc35fAV:-Kh6a6MFOGrTBIoYh5QY:bcsd6pc) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/engine.go#L26)):*
> ```Go
> func initEngine() error {
> 	//init Cache
> 	/*pendingQueries = &LRUCache{}
> ```

FYI: Reviewable does not deal well with this style of comment blocks.

---

*[rainsd/inbox.go, line 109 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6auY4TUGu-VmAXBtz:-Kh6auY4TUGu-VmAXBu-:b-2fmksa) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/inbox.go#L109)):*
> ```Go
> //processCapability processes capabilities and sends a notification back to the sender if the hash is not understood.
> func processCapability(caps []rainslib.Capability, sender ConnInfo, token rainslib.Token) {
> 	log.Info("Process capabilities")
> ```

This probably should be `log.Debug()`

---

*[rainsd/serverInfo.go, line 25 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6bETXrfk_Zd6J4jMK:-Kh6bETYUtiEdwLWRl6g:blccsqo) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInfo.go#L25)):*
> ```Go
> 	ServerPort          uint16
> 	MaxConnections      uint
> 	KeepAlivePeriodNano time.Duration
> ```

Since time.Duration does not have a fixed unit of measurement (though a resolution), you can drop `Nano`/`Micros` *unless* the parsing-from-config-file will assume that a unit-less integer is in ns.

Since I suspect you'll be using [Go's `time` module parser](https://golang.org/pkg/time/#ParseDuration), I think you should drop the units.

---

*[rainsd/serverInfo.go, line 118 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6cP2NpA2zeb6lXpe1:-Kh6cP2NpA2zeb6lXpe2:bn6dpw) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInfo.go#L118)):*
> ```Go
> 	//Returns false if the cache already contained an entry for the fourTuple.
> 	//If the cache is full it closes and removes a connection according to some metric
> 	Add(fourTuple string, conn net.Conn) bool
> ```

I think the first arg should be a struct like this (feel free to think of a better name):

```
typedef AddressPair struct {
 localIP net.IP
 localPort uint16
 remoteIP net.IP
 remotePort uint16
}
```

And then you can define a `String()` method on top that is used by the caches to make a hashable key. Users of the Cache would then just hand in `AddressPair`s. That way, the meaning of the keys is a bit more obvious. This is also a bit more consistent with the use of `ConnInfo` below.

---

*[rainsd/serverInfo.go, line 210 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6dlszFuG2GQuT4bUT:-Kh6dlszFuG2GQuT4bUU:b9s1mcq) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInfo.go#L210)):*
> ```Go
> 	section    *rainslib.AssertionSection
> 	validFrom  int64
> 	ValidUntil int64
> ```

Why is the visibility of the two validity members different? (uppercase/lowercase first letter)

---

*[rainsd/serverInfo.go, line 234 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6eOpOwXv_1TjzO-8A:-Kh6eOpOwXv_1TjzO-8B:bt44f4u) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInfo.go#L234)):*
> ```Go
> 	section    rainslib.MessageSectionWithSig
> 	validFrom  int64
> 	ValidUntil int64
> ```

Same visibility question as above.

---

*[rainsd/serverInterfaceImpl.go, line 310 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh76X1rxO1HwCDakbk-:-Kh76X1s3eD_L5UP8DtK:b-e4r6qo) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInterfaceImpl.go#L310)):*
> ```Go
> 
> //updateCount increases the element count by one and if it exceeds the cache size, deletes all sections from the least recently used cache entry.
> func updateCount(c *pendingSignatureCacheImpl) {
> ```

The name of this function is misleading since it's not directly obvious that it will expire old entries. I recommend either splitting of that functionality or renaming the function.

---

*[utils/assertionCache/assertionCache.go, line 28 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh7BVIM5qPKzutTd4Vp:-Kh7BVIM5qPKzutTd4Vq:b4o8p0t) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/utils/assertionCache/assertionCache.go#L28)):*
> ```Go
> //New creates a cache where the first parameter entry must contain the maximum size of the cache (>0).
> //The second Parameter specifies if cacheAnyContext is used. on input 'anyContext' is it used and on 'noAnyContext' not.
> func New(params ...interface{}) (*Cache, error) {
> ```

Add a check for the size==0 and then return a specific error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8)*

fehlmach:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[rainsd/engine.go, line 26 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6aYAWwriA9Sc35fAV:-Kh7f09lBVRU33wVnXSB:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/engine.go#L26)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

FYI: Reviewable does not deal well with this style of comment blocks.
</blockquote></details>

Done.

---

*[rainsd/inbox.go, line 109 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6auY4TUGu-VmAXBtz:-Kh7fERCWv1iAzYdnx-l:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/inbox.go#L109)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

This probably should be `log.Debug()`
</blockquote></details>

Done.

---

*[rainsd/serverInfo.go, line 25 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6bETXrfk_Zd6J4jMK:-Kh7fPTCRaiX0zqcDjfw:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInfo.go#L25)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Since time.Duration does not have a fixed unit of measurement (though a resolution), you can drop `Nano`/`Micros` *unless* the parsing-from-config-file will assume that a unit-less integer is in ns.

Since I suspect you'll be using [Go's `time` module parser](https://golang.org/pkg/time/#ParseDuration), I think you should drop the units.
</blockquote></details>

Done.

---

*[rainsd/serverInfo.go, line 118 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6cP2NpA2zeb6lXpe1:-Kh7krwZm2w0fqkfm1NC:bltana3) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInfo.go#L118)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

I think the first arg should be a struct like this (feel free to think of a better name):

```
typedef AddressPair struct {
 localIP net.IP
 localPort uint16
 remoteIP net.IP
 remotePort uint16
}
```

And then you can define a `String()` method on top that is used by the caches to make a hashable key. Users of the Cache would then just hand in `AddressPair`s. That way, the meaning of the keys is a bit more obvious. This is also a bit more consistent with the use of `ConnInfo` below.
</blockquote></details>

Done. I do not want to make it specially for IP, so I replaced it with 2 connInfo object. So it gets easier when we make it scion compatible.

---

*[rainsd/serverInfo.go, line 210 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6dlszFuG2GQuT4bUT:-Kh7lBjaLP8Xq8Qy_fEq:b-nrc84j) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInfo.go#L210)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Why is the visibility of the two validity members different? (uppercase/lowercase first letter)
</blockquote></details>

Done. typo

---

*[rainsd/serverInfo.go, line 234 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh6eOpOwXv_1TjzO-8A:-Kh7lLZKIeJZpvJSTWbw:b-m33llx) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInfo.go#L234)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Same visibility question as above.
</blockquote></details>

Done. copy past ;-)

---

*[rainsd/serverInterfaceImpl.go, line 310 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh76X1rxO1HwCDakbk-:-Kh7nm3m6Qc0wuHG5BmZ:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/rainsd/serverInterfaceImpl.go#L310)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

The name of this function is misleading since it's not directly obvious that it will expire old entries. I recommend either splitting of that functionality or renaming the function.
</blockquote></details>

Done.

---

*[utils/assertionCache/assertionCache.go, line 28 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-Kh7BVIM5qPKzutTd4Vp:-Kh7oTJpkCtOQml1pAdk:bd675tu) ([raw file](https://github.com/netsec-ethz/rains/blob/24a968ad37acddb07dca27576615bff0e8eca571/utils/assertionCache/assertionCache.go#L28)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Add a check for the size==0 and then return a specific error.
</blockquote></details>

Uhh, this class is deprecated. I forgot to delete it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8)*

fehlmach:
PendingQueryCache and negativeAssertionCache ready for review

---

Review status: 4 of 13 files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-:-Kh7pCRWaqPqXJHuSuYs:brumyqi)*

klausman:
Reviewed 9 of 9 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[rainsd/serverInterfaceImpl.go, line 493 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhLmwLp06ErgQgHdIpa:-KhLmwLq6zRerSX7BWPb:bf4eluz) ([raw file](https://github.com/netsec-ethz/rains/blob/7b3ffa77e0e8a207eb88b6bfac03b45f26e84b45/rainsd/serverInterfaceImpl.go#L493)):*
> ```Go
> 	//cache entry was deleted in the meantime. Retry
> 	log.Warn("Cache entry was delete between, trying to add new and getting the existing one. This case must be rare!")
> 	return c.Add(context, zone, name, objType, value)
> ```

I'm worried that this might recurse infinitely in the case of a bug elsewhere. How hard do you think it would be to limit this?

---

*[rainsd/serverInterfaceImpl.go, line 503 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhLnKsnAIwkbcc-L24l:-KhLnKsnAIwkbcc-L24m:b-dbn41k) ([raw file](https://github.com/netsec-ethz/rains/blob/7b3ffa77e0e8a207eb88b6bfac03b45f26e84b45/rainsd/serverInterfaceImpl.go#L503)):*
> ```Go
> }
> 
> //handlePendingQueryCacheSize deletes all sender infos from the least recently used cache entry if it exceeds the cache size
> ```

This assumes that we never exceed the cache size by more than one. Is there a scenario where we may have to delete multiplen entries?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8)*

fehlmach:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[rainsd/serverInterfaceImpl.go, line 493 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhLmwLp06ErgQgHdIpa:-KhLyKCQorhjZvBNBGFy:b-hnkovn) ([raw file](https://github.com/netsec-ethz/rains/blob/7b3ffa77e0e8a207eb88b6bfac03b45f26e84b45/rainsd/serverInterfaceImpl.go#L493)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

I'm worried that this might recurse infinitely in the case of a bug elsewhere. How hard do you think it would be to limit this?
</blockquote></details>

Not so hard, we just have to add a method getOrAdd() to the cache which directly returns a pointer to the set. Then we do not need to recurse. I am planning to add this and also an update functionality to the cache, after I have a running prototype of the server (meaning the engine.go is implemented)

---

*[rainsd/serverInterfaceImpl.go, line 503 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhLnKsnAIwkbcc-L24l:-KhLur3pMW9m-8AGT0QH:b-numyfk) ([raw file](https://github.com/netsec-ethz/rains/blob/7b3ffa77e0e8a207eb88b6bfac03b45f26e84b45/rainsd/serverInterfaceImpl.go#L503)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

This assumes that we never exceed the cache size by more than one. Is there a scenario where we may have to delete multiplen entries?
</blockquote></details>

It can happen, that multiple deletes get the same leastrecentlyusedKey() and then we might exceed the cache size for some amount of time. But this is not so problematic as we have a reap() function which periodically goes over the cache and deletes all expired values. Additionally, If the add method does not create a new cache entry (when there is already a query on the way), then we still delete a cache entry as the elementCount is larger than maxElements. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8)*

fehlmach:
I finished implementing the basic functionality of all caches that we need so far in the server.  After I will have implemented the engine and played around with it. I am going to solve the pending TODO's and start writing tests for all of them.

---

Review status: 8 of 13 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-:-KhLzHrmcJE4nWVZqFZm:b-azjk4p)*

klausman:
Reviewed 5 of 5 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8)*

klausman:
Reviewed 14 of 15 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[rainsd/consistencyChecks.go, line 112 at r4](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhbY3Wu7wrhx15L1l7Y:-KhbY3Wu7wrhx15L1l7Z:b-9fhk8a) ([raw file](https://github.com/netsec-ethz/rains/blob/194ba4f58020dfab9852e883766b5a1816c6610a/rainsd/consistencyChecks.go#L112)):*
> ```Go
> //dropAllWithContextZone deletes all assertions, shards and zones in the cache with the given context and zone
> func dropAllWithContextZone(context, zone string) {
> 	log.Warn("Inconsistency encountered. Drop all sections for given context and zone.", "context", context, "zone", zone)
> ```

I think this log message would better emitted at the points where this function is called, since the context (i.e. what the inconsistency was) is likely useful). If those spots already have logging, you can just drop this line.

---

*[rainsd/consistencyChecks.go, line 149 at r4](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhbYiCI7nwMF-RrVQV5:-KhbYiCI7nwMF-RrVQV6:b-8fwhjq) ([raw file](https://github.com/netsec-ethz/rains/blob/194ba4f58020dfab9852e883766b5a1816c6610a/rainsd/consistencyChecks.go#L149)):*
> ```Go
> 			}
> 		default:
> 			log.Warn(fmt.Sprintf("Not supported type. Expected *ShardSection or *AssertionSection. Got=%T", v))
> ```

In the case of an unsupported type, I don't think we ever want to return `true`. While I *think* the logic works out for most cases, an explicit `return false` here would remove all doubt.

---

*[rainsd/serverInterfaceImpl.go, line 1141 at r4](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhbcaQhT72dMpFGknv5:-KhbcaQhT72dMpFGknv6:blwmo2c) ([raw file](https://github.com/netsec-ethz/rains/blob/194ba4f58020dfab9852e883766b5a1816c6610a/rainsd/serverInterfaceImpl.go#L1141)):*
> ```Go
> 						ok := deleteAssertionFromRangeMap(c, v.section, v.validFrom, v.validUntil)
> 						if !ok {
> 							log.Error("Was not able to delete assertion from rangeMap")
> ```

Can we log some more useful info here? E.g. the assertion that could not be deleted? Otherwise, the logging is a bit useless.

---

*[rainslib/datamodel.go, line 378 at r4](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhbeKnjyO_WnldT_MjN:-KhbeKnjyO_WnldT_MjO:b82ier1) ([raw file](https://github.com/netsec-ethz/rains/blob/194ba4f58020dfab9852e883766b5a1816c6610a/rainslib/datamodel.go#L378)):*
> ```Go
> 	Context string
> 	Type    ObjectType
> 	Expires int64
> ```

Add a comment regarding the what this is (e.g. `//  Microseconds since the Unix epoch`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8)*

fehlmach:
Review status: 13 of 16 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[rainsd/consistencyChecks.go, line 112 at r4](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhbY3Wu7wrhx15L1l7Y:-Khzm33gOD63lPncnCmL:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/194ba4f58020dfab9852e883766b5a1816c6610a/rainsd/consistencyChecks.go#L112)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

I think this log message would better emitted at the points where this function is called, since the context (i.e. what the inconsistency was) is likely useful). If those spots already have logging, you can just drop this line.
</blockquote></details>

Done.

---

*[rainsd/consistencyChecks.go, line 149 at r4](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhbYiCI7nwMF-RrVQV5:-KhzmdA5KVPR87Uxx7Tz:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/194ba4f58020dfab9852e883766b5a1816c6610a/rainsd/consistencyChecks.go#L149)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

In the case of an unsupported type, I don't think we ever want to return `true`. While I *think* the logic works out for most cases, an explicit `return false` here would remove all doubt.
</blockquote></details>

Done.

---

*[rainsd/serverInterfaceImpl.go, line 1141 at r4](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhbcaQhT72dMpFGknv5:-KhznAH92qnIQFNwteqi:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/194ba4f58020dfab9852e883766b5a1816c6610a/rainsd/serverInterfaceImpl.go#L1141)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Can we log some more useful info here? E.g. the assertion that could not be deleted? Otherwise, the logging is a bit useless.
</blockquote></details>

Done.

---

*[rainslib/datamodel.go, line 378 at r4](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-KhbeKnjyO_WnldT_MjN:-Khzo9b_MjLD3qUb6MCq:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/194ba4f58020dfab9852e883766b5a1816c6610a/rainslib/datamodel.go#L378)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Add a comment regarding the what this is (e.g. `//  Microseconds since the Unix epoch`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8)*

klausman:
Reviewed 3 of 3 files at r5.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8)*

britram:
:lgtm:

---

Reviewed 2 of 13 files at r1, 11 of 15 files at r4, 3 of 3 files at r5.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-:-Ki-cuAaWp4Wn0PZXisM:bnfp4nl)*

britram:
:lgtm:

---

Reviewed 6 of 6 files at r6.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/8#-:-Ki-fsGLQF1FGukVQyli:bnfp4nl)*

PR #42, opened 2017-07-12
Design proposals for rains caches and their implementation

fehlmach:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/42)


britram:
Reviewed 64 of 64 files at r1.
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---

*[DESIGN-ASSERTION-CACHE.md, line 11 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KouysGYjkiU0e6rhn5Y:-KouysGYjkiU0e6rhn5Z:b-1ahjv9) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L11)):*
> ```Markdown
> ## Assertion requirements
> - cache has a fixed size which is configurable (to avoid memory exhaustion of the server in case of
>   an attack).
> ```

I'm not sure that it necessarily needs to be fixed-size, but definitely max-size -- whatever is simpler to implement.

---

*[DESIGN-ASSERTION-CACHE.md, line 13 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KouzICYbK3D0dhJqb0s:-KouzICYbK3D0dhJqb0t:b-o871no) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L13)):*
> ```Markdown
>   an attack).
> - In case the cache is full the least recently used assertion over which the server has no
>   authority is removed from the cache.
> ```

If the server has authority over more assertions than fit in the cache, complain VERY LOUDLY and drop the LRU or next-expiring authoritative assertion (whatever's easier, the main idea is to fail loudly, since a human / outer layer control loop will need to reconfigure)

---

*[DESIGN-ASSERTION-CACHE.md, line 22 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovAAoIFw6CnmANS4bK:-KovAAoJ4zZJdUdomWlj:bdcl12c) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L22)):*
> ```Markdown
>   can decide which entry it wants to send back according to a policy.
> - it must provide fast lookup of a set of assertions based on a zone, context and name interval.
>   This is necessary to check if a new shard or zone is consistent with the cached assertions. 
> ```

Is this correctly a function of the assertion cache?

---

*[DESIGN-ASSERTION-CACHE.md, line 23 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-Kov9Q1lOeI38_xj2FLL:-Kov9Q1lOeI38_xj2FLM:b-nyon1g) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L23)):*
> ```Markdown
> - it must provide fast lookup of a set of assertions based on a zone, context and name interval.
>   This is necessary to check if a new shard or zone is consistent with the cached assertions. 
> - it must provide a cleanup function that removes expired assertions.
> ```

I'd call this `reap()`. I'm not sure this should be part of the cache API, or perhaps it might simply be a no-op depending on implementation... Consider a fixed-size cache. Here you can be lazy, as there's actually no win to removing entries if the cache is really fixed size. What you really need is (1) to reject any assertion as invalid if it's expired when you retrieve it and (2) to choose the correct assertion to remove on an insertion to a full cache.

---

*[DESIGN-ASSERTION-CACHE.md, line 38 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovAZ1JsX13g6CBVh52:-KovAZ1JsX13g6CBVh53:b2eupcd) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L38)):*
> ```Markdown
> - to allow fast lookups of assertions for consistency checks. There is an additional hashmap keyed
>   by zone and context. The value points to a list of assertions ordered according to their name.
>   (Instead of an order list we could use a segment or interval tree) 
> ```

I'm starting to wonder whether consistency checking should best be supported by a separate data structure which maintains links into assertion and shard caches.

---

*[DESIGN-CONNECTION-CAPABILITY.md, line 22 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovL8IyCpJpr5XXt5go:-KovL8IzklvF1TiscdPq:b-xj5jd3) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-CONNECTION-CAPABILITY.md#L22)):*
> ```Markdown
>   a hash to capability mapping is removed from the cache, it is still clear what the capabilities
>   are. We store a pointer to reduce storage overhead.
> - The capability list of the own server and the hash thereof are loaded from the config file.
> ```

hash should be calculated at config load time, no?

---

*[DESIGN-CONNECTION-CAPABILITY.md, line 33 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovLHvAohdQCwtbNk9c:-KovLHvAohdQCwtbNk9d:b-st1hba) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-CONNECTION-CAPABILITY.md#L33)):*
> ```Markdown
> - all cache operations must be safe for concurrent access
> 
> ## zone key cache implementation
> ```

zone key? why is this here?

---

*[DESIGN-CONNECTION-CAPABILITY.md, line 50 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovLO8t_sYDOuBZYm6r:-KovLO8t_sYDOuBZYm6s:b-cmn1aj) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-CONNECTION-CAPABILITY.md#L50)):*
> ```Markdown
> - it must provide fast lookup to connections and pointers to capability lists based on the
>   connection's type and addr. If there are several connections stored, it returns all of them.
> - it must provide a mechanism to detect and delete closed connections.
> ```

closure is usually noticed on a read or write attempt after close. not sure how `select()` style IO works in Go.

---

*[DESIGN-NEG-CACHE.md, line 5 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovDE9hzKGkWSoCTtY5:-KovDE9hzKGkWSoCTtY6:b9pbwjg) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L5)):*
> ```Markdown
> ## Cache design decisions
> - Shards and Zones over which the server has authority are only removed when they expire. All other
>   shards and zones are subject to a least recently used policy.
> ```

see assertion cache design for what happens when you hit max on authority.

---

*[DESIGN-NEG-CACHE.md, line 7 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovDObkuuIpGhySpw3y:-KovDObkuuIpGhySpw3z:b-k4n0hd) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L7)):*
> ```Markdown
>   shards and zones are subject to a least recently used policy.
> - Expired shards and zones are also returned such that it is possible to answer a query which has
>   query option 5 (expired assertions are acceptable) set. 
> ```

Not necessary to design to make option 5 useful.

---

*[DESIGN-NEG-CACHE.md, line 13 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovEuOacrRfUsZN90jL:-KovEuOacrRfUsZN90jM:bnbnp0a) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L13)):*
> ```Markdown
>   an attack).
> - In case the cache is full the least recently used shard or zone over which the server has no
>   authority is removed from the cache.
> ```

see assertion cache design for notes on "fixed size"-ness

---

*[DESIGN-NEG-CACHE.md, line 22 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovH6-otsz9o02CvCxb:-KovH6-pmI3UonCJ79ME:b-z92as4) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L22)):*
> ```Markdown
>   consistency checks between a new shard and all overlapping cached shards and zones. A set is
>   returned such that the calling function can decide which entry it wants to send back according to
>   a policy
> ```

not sure I follow here -- aren't we missing the standard case where we have a name and want a neg for it? see previous note that maybe a separate consistency checking cache is a good idea...

---

*[DESIGN-NEG-CACHE.md, line 23 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovH2qyEM2wwMbGUnGq:-KovH2qyEM2wwMbGUnGr:bhec3gz) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L23)):*
> ```Markdown
>   returned such that the calling function can decide which entry it wants to send back according to
>   a policy
> - it must provide a cleanup function that removes expired entries.
> ```

see assertion cache design on this point.

---

*[DESIGN-PENDING-QUERY-CACHE.md, line 18 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovMgVqInYMfOMOsEK5:-KovMgVrxFPo8dT9FanN:bvj16vv) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-PENDING-QUERY-CACHE.md#L18)):*
> ```Markdown
> ## Pending query cache implementation proposal
> 1. We only remove an element from the cache after we have received an answer or the query has
>    expired. In this case the server must be able to decide the query expiration time it sends to not
> ```

A RAINS server might decide to expire a query earlier than its expiration time, and send an early 504.

---

*[DESIGN-PENDING-QUERY-CACHE.md, line 28 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovN-Rd35EXTvrgBY4S:-KovN-Rd35EXTvrgBY4T:b-8813hm) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-PENDING-QUERY-CACHE.md#L28)):*
> ```Markdown
>    only queries for this authority with a large expiration time that it does not keep these entries
>    for a long time and drop all other queries in the meantime.
> 2. We remove the least recently used query from the cache in case it is full. This has the
> ```

Again, if we decide all caches should be fixed-size, this is probably the way to go.

---

*[DESIGN-PENDING-SIGNATURE-CACHE.md, line 21 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovLcNcwLlQYwj1JPmg:-KovLcNdPby5EOL8PCXm:bthv4ak) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-PENDING-SIGNATURE-CACHE.md#L21)):*
> ```Markdown
>   public keys can simultaneously be issued by this server. If the active Token cache is full and the
>   section needs a not yet queried public key it gets dropped even when there is space in the pending
>   signature cache.
> ```

Token cache and pending signature cache overruns should result in VERY LOUD complaints. Realistically, this should only happen in an active DoS situation. DoS defense might include changes to cache expiration policy.

---

*[DESIGN-REV-LOOKUP-CACHE.md, line 19 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovIxLjc2PFgn4zl5_g:-KovIxLkzrNzEIYZdrsl:bv0xhem) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-REV-LOOKUP-CACHE.md#L19)):*
> ```Markdown
> 1. There are two binary tries per context (one for IP4 and the other for IP6 reverse (name) lookup).
>    We then also need a map keyed by context, where the value points to two trie roots. Each existing
>    node of a trie stores a set of addressAssertions and addressZones.
> ```

contexts will, I think, be rare... but usually within a context you'll only have a tiny subset of the whole tree (i.e., a 1918 tree or ULA), so this proposal is probably the better way to do it.

---

*[DESIGN-REV-LOOKUP-CACHE.md, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovJCZG_h6dGG6pAmeV:-KovJCZG_h6dGG6pAmeW:b-ikz460) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-REV-LOOKUP-CACHE.md#L30)):*
> ```Markdown
>   because the trie entries are not rearranged so if you lock a node and all nodes in the subtree are
>   not locked by an other process one can safely delete or modify anything in the subtree by only
>   holding the lock on the root.) 
> ```

operationally, deletions will be rare, so requiring a trie root lock is fine.

---

*[DESIGN-REV-LOOKUP-CACHE.md, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovJLCPHMceGbG3snPc:-KovJLCQ11ovoBACwv-Q:b-l0tmqc) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-REV-LOOKUP-CACHE.md#L39)):*
> ```Markdown
> - Binary tries are used to quickly find the longest prefix match.
> - The tries are built dynamically (a full ipv6 trie is too large to be stored. It has more than
>   2^128 entries. In a real setting it is sparse though).
> ```

It might make sense for the generic trie structure to allow bits to be taken k at a time (for 2^k children per node), in order to account for v6 sparseness.

---

*[DESIGN-TOKEN-CACHE.md, line 19 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovJrED2rw2MaSY-6lM:-KovJrEEt-wJee_685T_:bjcu9az) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-TOKEN-CACHE.md#L19)):*
> ```Markdown
> - it must provide a fast response to the question if a token is contained in the cache.
> - it must provide a cleanup function that removes expired entries and logs information about the
>   entity which did not respond (What information should we log such that we blame the right entity?)
> ```

for outgoing queries, we should log who we sent the query to.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42)*

fehlmach:
Review status: all files reviewed at latest revision, 21 unresolved discussions.

---

*[DESIGN-ASSERTION-CACHE.md, line 8 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovhD5v9LKVlLFQhc0W:-KovhD5v9LKVlLFQhc0X:bi4f1uh) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L8)):*
> ```Markdown
> - Expired assertions are also returned such that it is possible to answer a query which has
>   query option 5 (expired assertions are acceptable) set. 
>   
> ```

Store delegation assertions not in the assertion cache but in the zoneKeyCache. So delegations cannot be pushed out of the assertion cache by non delegation assertions. The size of the zone key cache should be chosen in a way that it only evicts an entry on expiration. A lru node contains all delegation assertions of a zone and context. So in case the maximum cache size is hit we remove all delegation keys of a zone and context or none. Thus, we do not end up in a situation where a client sends a delegation query and obtains as a result some delegation assertions where those might not match the current key phase. By issuing the same query we again would get the same answer. we would be stuck.  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42)*

fehlmach:
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---

*[DESIGN-ASSERTION-CACHE.md, line 11 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KouysGYjkiU0e6rhn5Y:-Kp0Pvfh7hR7_g3uwjbZ:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L11)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

I'm not sure that it necessarily needs to be fixed-size, but definitely max-size -- whatever is simpler to implement.
</blockquote></details>

Done.

---

*[DESIGN-ASSERTION-CACHE.md, line 13 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KouzICYbK3D0dhJqb0s:-Kp0Q_jR7-K5gpUakVGs:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L13)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

If the server has authority over more assertions than fit in the cache, complain VERY LOUDLY and drop the LRU or next-expiring authoritative assertion (whatever's easier, the main idea is to fail loudly, since a human / outer layer control loop will need to reconfigure)
</blockquote></details>

Done.

---

*[DESIGN-ASSERTION-CACHE.md, line 23 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-Kov9Q1lOeI38_xj2FLL:-Kp0RgjHNVY66syWKq3d:b-e7qepx) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L23)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

I'd call this `reap()`. I'm not sure this should be part of the cache API, or perhaps it might simply be a no-op depending on implementation... Consider a fixed-size cache. Here you can be lazy, as there's actually no win to removing entries if the cache is really fixed size. What you really need is (1) to reject any assertion as invalid if it's expired when you retrieve it and (2) to choose the correct assertion to remove on an insertion to a full cache.
</blockquote></details>

So the lookup function should have a flag if expired assertions should be returned? To be able to answer queries with opt 5 enabled (accept expired entries)

---

*[DESIGN-ASSERTION-CACHE.md, line 38 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovAZ1JsX13g6CBVh52:-Kp0S1lGsSR5Y7-jMqeU:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L38)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

I'm starting to wonder whether consistency checking should best be supported by a separate data structure which maintains links into assertion and shard caches.
</blockquote></details>

Done.

---

*[DESIGN-CONNECTION-CAPABILITY.md, line 22 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovL8IyCpJpr5XXt5go:-Kp0S7m0B7-kn2AsY3IZ:b2nhswv) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-CONNECTION-CAPABILITY.md#L22)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

hash should be calculated at config load time, no?
</blockquote></details>

Yep

---

*[DESIGN-CONNECTION-CAPABILITY.md, line 33 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovLHvAohdQCwtbNk9c:-Kp0SQa9KxgwQ7VIITCh:b-8ixk8h) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-CONNECTION-CAPABILITY.md#L33)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

zone key? why is this here?
</blockquote></details>

copy&paste error

---

*[DESIGN-CONNECTION-CAPABILITY.md, line 50 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovLO8t_sYDOuBZYm6r:-Kp0TZnOP6MU5u3P-qhP:b-kmqqki) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-CONNECTION-CAPABILITY.md#L50)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

closure is usually noticed on a read or write attempt after close. not sure how `select()` style IO works in Go.
</blockquote></details>

Cache does not remove closed connections. If the process writing to the connection notices that it was closed, it establishes a new connection and updates the cache entry 

---

*[DESIGN-NEG-CACHE.md, line 5 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovDE9hzKGkWSoCTtY5:-Kp0UJFZ6DFsabImuVoN:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L5)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

see assertion cache design for what happens when you hit max on authority.
</blockquote></details>

Done.

---

*[DESIGN-NEG-CACHE.md, line 7 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovDObkuuIpGhySpw3y:-Kp0ULj3OJUFLWLNWySk:b3mm8fa) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L7)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

Not necessary to design to make option 5 useful.
</blockquote></details>

see question at assertion cache design for the same issue

---

*[DESIGN-NEG-CACHE.md, line 13 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovEuOacrRfUsZN90jL:-Kp0Ug0aQwCFipahSI8I:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L13)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

see assertion cache design for notes on "fixed size"-ness
</blockquote></details>

Done.

---

*[DESIGN-NEG-CACHE.md, line 23 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovH2qyEM2wwMbGUnGq:-Kp0UpIpS_Q0YUy3t9Hb:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L23)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

see assertion cache design on this point.
</blockquote></details>

Done.

---

*[DESIGN-PENDING-QUERY-CACHE.md, line 18 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovMgVqInYMfOMOsEK5:-Kp0jGFhBxtPPIzWSHQC:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-PENDING-QUERY-CACHE.md#L18)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

A RAINS server might decide to expire a query earlier than its expiration time, and send an early 504.
</blockquote></details>

Done.

---

*[DESIGN-PENDING-QUERY-CACHE.md, line 28 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovN-Rd35EXTvrgBY4S:-Kp0emN8eAq_HghbDV5C:bbaylr3) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-PENDING-QUERY-CACHE.md#L28)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

Again, if we decide all caches should be fixed-size, this is probably the way to go.
</blockquote></details>

wrong terminology. most of them are now maximum size caches.

---

*[DESIGN-PENDING-SIGNATURE-CACHE.md, line 21 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovLcNcwLlQYwj1JPmg:-Kp0ZFBaYBrDFoa_MphK:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-PENDING-SIGNATURE-CACHE.md#L21)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

Token cache and pending signature cache overruns should result in VERY LOUD complaints. Realistically, this should only happen in an active DoS situation. DoS defense might include changes to cache expiration policy.
</blockquote></details>

Done.

---

*[DESIGN-TOKEN-CACHE.md, line 19 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovJrED2rw2MaSY-6lM:-Kp0_1suIHZFs9GpOPGd:b4ghr6u) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-TOKEN-CACHE.md#L19)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

for outgoing queries, we should log who we sent the query to.
</blockquote></details>

This is done. -> see token cache implementation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42)*

fehlmach:
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---

*[DESIGN-ASSERTION-CACHE.md, line 22 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovAAoIFw6CnmANS4bK:-KpFNT9_bVRjMshcZgPd:b68m9o6) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L22)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

Is this correctly a function of the assertion cache?
</blockquote></details>

moved to the consistency cache.

---

*[DESIGN-NEG-CACHE.md, line 22 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovH6-otsz9o02CvCxb:-KpFNYJtg0yNYo5-ZVb2:b-2ip9jv) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-NEG-CACHE.md#L22)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

not sure I follow here -- aren't we missing the standard case where we have a name and want a neg for it? see previous note that maybe a separate consistency checking cache is a good idea...
</blockquote></details>

yep, introduced now a new consistency cache.

---

*[DESIGN-REV-LOOKUP-CACHE.md, line 19 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovIxLjc2PFgn4zl5_g:-KpFU95GpOOqPh7_UGPV:b-7oj6nk) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-REV-LOOKUP-CACHE.md#L19)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

contexts will, I think, be rare... but usually within a context you'll only have a tiny subset of the whole tree (i.e., a 1918 tree or ULA), so this proposal is probably the better way to do it.
</blockquote></details>

jap.

---

*[DESIGN-REV-LOOKUP-CACHE.md, line 30 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovJCZG_h6dGG6pAmeV:-KpFUIgZXgruA2jMMnqq:b32wtd) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-REV-LOOKUP-CACHE.md#L30)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

operationally, deletions will be rare, so requiring a trie root lock is fine.
</blockquote></details>

ok

---

*[DESIGN-REV-LOOKUP-CACHE.md, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KovJLCPHMceGbG3snPc:-KpFUL2dtTwwr0iQZv-H:b-3rvi06) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-REV-LOOKUP-CACHE.md#L39)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

It might make sense for the generic trie structure to allow bits to be taken k at a time (for 2^k children per node), in order to account for v6 sparseness.
</blockquote></details>

jap makes sense!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42)*

klausman:
Reviewed 59 of 64 files at r1, 41 of 41 files at r2.
Review status: all files reviewed at latest revision, 24 unresolved discussions.

---

*[DESIGN-ASSERTION-CACHE.md, line 8 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KorfiXwwSwV4ey77I8R:-KorfiXwwSwV4ey77I8S:bdhkr2z) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L8)):*
> ```Markdown
> - Expired assertions are also returned such that it is possible to answer a query which has
>   query option 5 (expired assertions are acceptable) set. 
>   
> ```

Trailing WS

---

*[DESIGN-ASSERTION-CACHE.md, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KorgGH0cJufRO6E2jmO:-KorgGH0cJufRO6E2jmP:bhy8frc) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L39)):*
> ```Markdown
>   by zone and context. The value points to a list of assertions ordered according to their name.
>   (Instead of an order list we could use a segment or interval tree) 
> - sections over which this server has authority are not subject to lru removal.
> ```

Do they still reside in the cache? or are authoritative sections kept in a (somewhat) static data structure?

---

*[DESIGN-CONSISTENCY-CACHE.md, line 8 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KpPYtfx6NXXzLmimM5X:-KpPYtfx6NXXzLmimM5Y:bdhkr2z) ([raw file](https://github.com/netsec-ethz/rains/blob/59af3ed0b81f1f65d95d322f9d3eea70f716a245/DESIGN-CONSISTENCY-CACHE.md#L8)):*
> ```Markdown
>   checks in parallel to answering queries as the write lock is not on the assertion or negative
>   assertion cache (which increases efficiency).
>   
> ```

Trailing WS

---

*[DESIGN-CONSISTENCY-CACHE.md, line 16 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KpPZ0hV1gFt_HmBM2gs:-KpPZ0hV1gFt_HmBM2gt:b17dtj3) ([raw file](https://github.com/netsec-ethz/rains/blob/59af3ed0b81f1f65d95d322f9d3eea70f716a245/DESIGN-CONSISTENCY-CACHE.md#L16)):*
> ```Markdown
> - Every element in the assertion and negative assertion cache must also be present in the
>   consistency cache and vice versa. The assertion and negative assertion caches are responsible to
>   remove the corresponding entry from the consistency cache if they evict an entry in their cache.  
> ```

Trailing WS (won't point out further occurences).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42)*

fehlmach:
Review status: all files reviewed at latest revision, 24 unresolved discussions.

---

*[DESIGN-ASSERTION-CACHE.md, line 8 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KorfiXwwSwV4ey77I8R:-KpYmXessK5FYKQkEDIh:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L8)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Trailing WS
</blockquote></details>

Done.

---

*[DESIGN-ASSERTION-CACHE.md, line 39 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KorgGH0cJufRO6E2jmO:-KpYmbxE2Yq-bOlOgM_Y:b-63zjtx) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L39)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Do they still reside in the cache? or are authoritative sections kept in a (somewhat) static data structure?
</blockquote></details>

They stay in the cache until they are removed by the reap function after they have been expired. 

---

*[DESIGN-CONSISTENCY-CACHE.md, line 8 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KpPYtfx6NXXzLmimM5X:-KpYn38zV03q4EQ9YcEi:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/59af3ed0b81f1f65d95d322f9d3eea70f716a245/DESIGN-CONSISTENCY-CACHE.md#L8)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Trailing WS
</blockquote></details>

Done.

---

*[DESIGN-CONSISTENCY-CACHE.md, line 16 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-KpPZ0hV1gFt_HmBM2gs:-KpYn3sT_Tj5nKN7QpHh:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/59af3ed0b81f1f65d95d322f9d3eea70f716a245/DESIGN-CONSISTENCY-CACHE.md#L16)):*
<details><summary><i>Previously, klausman (Tobias Klausmann) wrote…</i></summary><blockquote>

Trailing WS (won't point out further occurences).
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42)*

klausman:
Reviewed 3 of 64 files at r1, 9 of 9 files at r3.
Review status: all files reviewed at latest revision, 20 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42)*

britram:
Reviewed 32 of 41 files at r2, 9 of 9 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[DESIGN-ASSERTION-CACHE.md, line 23 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-Kov9Q1lOeI38_xj2FLL:-KpsSbmtire4KLn_UpOb:bz8zkf0) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L23)):*
<details><summary><i>Previously, fehlmach wrote…</i></summary><blockquote>

So the lookup function should have a flag if expired assertions should be returned? To be able to answer queries with opt 5 enabled (accept expired entries)
</blockquote></details>

If it weren't necessary to handle in the interface, this would really be a sub-design-level detail. Two ways to do this: make the cache interface contract return everything, including not-yet-reaped but expired queries, and make it the caller's responsibility to handle option 5, or put option 5 directly into the interface. I have a slight preference for the former, since it makes the cache a cache and doesn't push vagaries of the protocol into the cache design.

Note that option 5 doesn't mean "please return expired entries", it means "okay to return expired entries". This option is only useful together with "don't tell anyone I asked", and could possibly folded into it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42)*

britram:
:lgtm:

---

Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-:-KpsW5kxSvm_N9PSuEWd:bnfp4nl)*

fehlmach:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[DESIGN-ASSERTION-CACHE.md, line 23 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/42#-Kov9Q1lOeI38_xj2FLL:-KpsYhRDXjvCR7sx5M2X:b-10x2bc) ([raw file](https://github.com/netsec-ethz/rains/blob/5230f5fd9a34435d3b2dc6a8cb5c68bd1d3568be/DESIGN-ASSERTION-CACHE.md#L23)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

If it weren't necessary to handle in the interface, this would really be a sub-design-level detail. Two ways to do this: make the cache interface contract return everything, including not-yet-reaped but expired queries, and make it the caller's responsibility to handle option 5, or put option 5 directly into the interface. I have a slight preference for the former, since it makes the cache a cache and doesn't push vagaries of the protocol into the cache design.

Note that option 5 doesn't mean "please return expired entries", it means "okay to return expired entries". This option is only useful together with "don't tell anyone I asked", and could possibly folded into it.
</blockquote></details>

Yes that makes sense. I also prefer option 1. I changed it in this way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/42)*

PR #44, opened 2017-07-31
Design documents and cache reimplementation

fehlmach:
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/44)


britram:
Reviewed 7 of 102 files at r1.
Review status: 7 of 87 files reviewed at latest revision, 20 unresolved discussions.

---

*[DESIGN-HANDLE-QUERIES.md, line 5 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNB1KB0dghOKvIVjc4:-KqNB1KB0dghOKvIVjc5:bjj512e) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L5)):*
> ```Markdown
> ## Design decisions
> - Delegation queries are handled slightly differently because they are important for the system to
>   run with few disruption.
> ```

s/few/minimal/

---

*[DESIGN-HANDLE-QUERIES.md, line 16 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNDoXRaJgj8BUI2BEd:-KqNDoXSbqrSgGjwK8IA:b-mk4l08) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L16)):*
> ```Markdown
>   queried server. In contrast, the sender of the section must have had all keys necessary to do the
>   verification. [editors note: update missing public key and cache design in case we decide to do
>   this case. If not, update this document]
> ```

cf. our discussion on Monday 31.7, I think this is obsolete.

---

*[DESIGN-HANDLE-QUERIES.md, line 19 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNBB1xs98cr7qO3fYM:-KqNBB1xs98cr7qO3fYN:b-vxhlbw) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L19)):*
> ```Markdown
> - Note that if a rains server decides to do the lookup by himself for a delegation query it might be
>   blacklisted in case the server to which it starts the lookup is not responding and thus, it can
>   also not respond in time.
> ```

cf. all the other notes on blacklisting in this review. The server should defend itself holistically; caches should not defend themselves in isolation.

---

*[DESIGN-HANDLE-QUERIES.md, line 40 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNEQGXPNc9EIKJoUKL:-KqNEQGYA3l4cj1uZVeC:b-a2k32b) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L40)):*
> ```Markdown
> ### Handle delegation queries
> 1. If the queried delegation is complete in the zoneKeyCache answer with all delegations of the
>    queried zone.
> ```

should this go through the zoneKeyCache, or should we make sure all delegation assertions stay live in the assertionCache? Would be nice if query handling weren't special-cased.

---

*[DESIGN-HANDLE-QUERIES.md, line 48 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNGM7YPK2lXZ1GDrsS:-KqNGM7YPK2lXZ1GDrsT:b7fu8dv) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L48)):*
> ```Markdown
>      assertions, else it responds with the redirect assertion to the zone. In case there is no
>      redirect assertion for the queried zone in the cache, it responds with a redirect to a root
>      server.
> ```

Whether a server sends redirects or follows them should be at least a global configuration option. See future redirect cache design document.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 6 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN2uM3aCq2qW4Xy25b:-KqN2uM4crgqlTRCVpLO:b-gq5p4g) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L6)):*
> ```Markdown
> - The callback function for the pending sections is not triggered by delegation assertions which can
>   be used to verify the signature but are not a response to the query issued for this information.
>   The event that another server pushes such a delegation assertion exactly in the moment where this
> ```

s/event/probability/

---

*[DESIGN-MISSING-PUBLICKEY.md, line 9 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN3-9Vp8DGhbRZOXrq:-KqN3-9Vp8DGhbRZOXrr:b-kch0vz) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L9)):*
> ```Markdown
>   server is waiting for this delegation assertion is negligible. But we would have to check each
>   incoming section if it matches any of the sections in the pending caches which incurs unnecessary
>   overhead.
> ```

What's the complexity of the check here? This seems a little like premature optimization to me.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 12 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN3KMViUYY6ePIbOOw:-KqN3KMWxVtV63sHr9XR:b-zoufs) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L12)):*
> ```Markdown
> - We do not keep state how many times we issued a query for a given zone. We repeat it until the
>   query expires (which is configurable). A malicious entity could send alternating redirects from
>   two of its servers and without an expiration for the query the server would ask them infinitely.
> ```

The server should *always* have a maximum validity for incoming queries -- the actual expiration is the min of the enforced max validity and the requested expiration. Indeed, this is a good general design pattern against state exhaustion.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 21 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN55_wcpi5V1dFTG4u:-KqN55_wcpi5V1dFTG4v:b-goa0f7) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L21)):*
> ```Markdown
>   has two or several rains servers, it is still possible to configure them such that only some of
>   them are doing recursive lookup. Those servers doing recursive lookup must be white-listed in the
>   local blacklisting service.
> ```

In general, speculation on policy at blacklisting servers can IMO be deferred. All we really need now is the ability to account for the provenance of a message, a way to log potentially suspicious messages with that provenance, and a way to blacklist servers and zones. Everything else is at this point future work.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 22 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN7aADZSwo5rk0ySwj:-KqN7aAEGearM6AvMHhh:b-8959x9) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L22)):*
> ```Markdown
>   them are doing recursive lookup. Those servers doing recursive lookup must be white-listed in the
>   local blacklisting service.
> - The communication partner is either an other rains server or rainspub. If the public key is
> ```

s/an other/another/ 

---

*[DESIGN-MISSING-PUBLICKEY.md, line 24 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN7eZlDJL_Fqhympq6:-KqN7eZlDJL_Fqhympq7:b-a5ujx1) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L24)):*
> ```Markdown
> - The communication partner is either an other rains server or rainspub. If the public key is
>   missing the server first has to determine where it should send a delegation query to. One approach
>   is to always send the delegation query to the entity that has sent the section. For a rains server
> ```

FYI I'll add that a server should always be prepared to send its keys to the rains doc.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 27 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN8ZHyto3_gbkd7fTL:-KqN8ZHyto3_gbkd7fTM:b-shnnwr) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L27)):*
> ```Markdown
>   this should in most cases result in a positive answer as the previous server also had to verify
>   the signatures. But if the section has come from rainspub then the server will get no answer or
>   maybe a notification message (501 server not capable or maybe a new notification type. Should it
> ```

I don't understand the problem here -- note that rainspub is not a server, it's a third kind of entity that can push data at a server. We can simply define that it MUST also provide the delegation assertion (and therefore the keys) for the assertions it's pushing.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN91MropzI-aZYvHgM:-KqN91MropzI-aZYvHgN:b6x5wc3) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L32)):*
> ```Markdown
>   expiration approach then we need to change the cache such that the reap function returns
>   information what and where we have to send the delegation query). [I would prefer the explicit
>   approach where rainspub sends a notification message back. Because it keeps the cache simple and
> ```

what? no. rainspub is one-way.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 44 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN9R7vMAgL-1PA3RxB:-KqN9R7wva6ZAQvxER2f:b6nqit1) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L44)):*
> ```Markdown
>    key(s))
> 2. Adds the section together with the token used on the previously sent delegation query and the
>    destination address to the pending signature cache. This step allows the handling go routine to
> ```

s/go routine/goroutine/

---

*[DESIGN-MISSING-PUBLICKEY.md, line 57 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN9vU-NL1WCrGxcQux:-KqN9b3tmUjcPp2s9kbG:b-fglx9x) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L57)):*
> ```Markdown
>     \+ Use the same process, no special case, less complexity
>     \- Slightly slower as the sections are not directly handled but added at the end of the queue.
>        As the queue should mostly contain few elements, the added latency is expected to be small.
> ```

yes! to avoiding unnecessary special cases.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 70 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNAAhZ7cmisHCU9dQ9:-KqNAAhZ7cmisHCU9dQA:b2dtxuk) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L70)):*
> ```Markdown
>   respond with all delegations they have for a given zone.)
> - Assertion(s) not all containing either a public key, an IP4, an IP6 or a redirect object for the
>   queried zone will be logged and possibly blacklisted. We do not want to allow normal assertion(s)
> ```

remember, blacklisting is a separate process.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 83 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNANHKLoqpx-VUutHY:-KqNANHKLoqpx-VUutHZ:b-k9b0sk) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L83)):*
> ```Markdown
>      the cache. Send a response back e.g. a notification?
>   2. Remove all other delegation assertions from the zoneKeyCache. Resend the query and update the
>      token on the pending signature cache.
> ```

a bad signature on a delegation isn't necessarily an inconsistency event

---

*[DESIGN-MISSING-PUBLICKEY.md, line 84 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNAfzzrFwCqY0QmwFw:-KqNAfzzrFwCqY0QmwFx:btukon5) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L84)):*
> ```Markdown
>   2. Remove all other delegation assertions from the zoneKeyCache. Resend the query and update the
>      token on the pending signature cache.
> - Notification Message:
> ```

...more detail here please.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 85 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNAZDeJjHZhicD9SsH:-KqNAZDeJjHZhicD9SsI:b-xj9qj1) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L85)):*
> ```Markdown
>      token on the pending signature cache.
> - Notification Message:
>   Depending on the notification type different behavior. Resend query, drop pending sections silent,
> ```

s/silent/silently.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNAuwhgzS2xbklPDYB:-KqNAuwhgzS2xbklPDYC:bst7phj) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L87)):*
> ```Markdown
>   Depending on the notification type different behavior. Resend query, drop pending sections silent,
>   send response to dropped pending sections
> - Assertion containing one redirect and one or several IP4/IP6 objects for the redirect name:
> ```

this logic should appear in the section in the redirect cache, I think...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/44)*

fehlmach:
Review status: 7 of 88 files reviewed at latest revision, 21 unresolved discussions.

---

*[DESIGN-HANDLE-QUERIES.md, line 5 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNB1KB0dghOKvIVjc4:-KqNXK-whi2MnlhJr0gl:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L5)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

s/few/minimal/
</blockquote></details>

Done.

---

*[DESIGN-HANDLE-QUERIES.md, line 16 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNDoXRaJgj8BUI2BEd:-KqNXVI-UBG8mGb3zypj:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L16)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

cf. our discussion on Monday 31.7, I think this is obsolete.
</blockquote></details>

Done.

---

*[DESIGN-HANDLE-QUERIES.md, line 19 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNBB1xs98cr7qO3fYM:-KqNY3Y0HkJpmdjmzbB_:b-qtkx4r) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L19)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

cf. all the other notes on blacklisting in this review. The server should defend itself holistically; caches should not defend themselves in isolation.
</blockquote></details>

This note does not concern the cache but the handling of a delegation query in general. We decided to log in the active token cache whenever an entry expires. Based on that an external service can then decide if it wants to blacklist this server. And I note here that if the server decides to do the lookup by himself it might miss this deadline and might get blacklisted by the external service.

---

*[DESIGN-HANDLE-QUERIES.md, line 40 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNEQGXPNc9EIKJoUKL:-KqNZD2tcx5dg-rVnGej:b-7r7ggz) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L40)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

should this go through the zoneKeyCache, or should we make sure all delegation assertions stay live in the assertionCache? Would be nice if query handling weren't special-cased.
</blockquote></details>

We could mark delegation assertions as internal in the assertion cache, then they get only evicted when they expire or the zone has sent an inconsistent message (in which case all sections from this zone are removed) What do you think about that? then we can handle them the same as any other query.

---

*[DESIGN-HANDLE-QUERIES.md, line 48 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNGM7YPK2lXZ1GDrsS:-KqN_WDSqth5NkitoiCt:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L48)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

Whether a server sends redirects or follows them should be at least a global configuration option. See future redirect cache design document.
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 6 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN2uM3aCq2qW4Xy25b:-KqNghMfrQ9xK1uGZcip:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L6)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

s/event/probability/
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 9 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN3-9Vp8DGhbRZOXrq:-KqNjEhgcNUdAAf4_j50:b-xmq2vv) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L9)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

What's the complexity of the check here? This seems a little like premature optimization to me.
</blockquote></details>

The main reason was that I wanted to wait until all delegation assertions came in before we answer the query. But as we now answer as soon as we have any information, we can remove this part. Concerning complexity, we can design the pending signature cache such that it takes one hashmap lookup + cleanup of activeToken cache.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 12 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN3KMViUYY6ePIbOOw:-KqNmOmHT7RG7RJqm2wU:b-4cknzz) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L12)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

The server should *always* have a maximum validity for incoming queries -- the actual expiration is the min of the enforced max validity and the requested expiration. Indeed, this is a good general design pattern against state exhaustion.
</blockquote></details>

This is what I was trying to say in this paragraph but I chose bad terminology. I rephrased it in a way such that it should be more clear.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 21 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN55_wcpi5V1dFTG4u:-KqNnNpKUwfS7JRuJBYM:b-83od3n) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L21)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

In general, speculation on policy at blacklisting servers can IMO be deferred. All we really need now is the ability to account for the provenance of a message, a way to log potentially suspicious messages with that provenance, and a way to blacklist servers and zones. Everything else is at this point future work.
</blockquote></details>

Noted

---

*[DESIGN-MISSING-PUBLICKEY.md, line 22 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN7aADZSwo5rk0ySwj:-KqNmooPALMRe_LmwQEC:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L22)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

s/an other/another/ 
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 24 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN7eZlDJL_Fqhympq6:-KqNn1Tox1JVenXeQLYD:byyr852) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L24)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

FYI I'll add that a server should always be prepared to send its keys to the rains doc.
</blockquote></details>

What is the rains doc?

---

*[DESIGN-MISSING-PUBLICKEY.md, line 27 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN8ZHyto3_gbkd7fTL:-KqNosgDWSKcHnEeyujm:b-o827b4) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L27)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

I don't understand the problem here -- note that rainspub is not a server, it's a third kind of entity that can push data at a server. We can simply define that it MUST also provide the delegation assertion (and therefore the keys) for the assertions it's pushing.
</blockquote></details>

I also though about this approach. The reason why I did not choose it was that rainspub must get this information beforehand. out-of-band or by a client run at the authority to query the super ordinate zone for the delegation.  In the case I described everything works in-band. Instead of sending something back to the rainspub we could just log an error message that the received content from rainspub could not be processed. We should IMO also raise an alarm such that an operator can check what the problem is before the zone gets unavailable. The alarm approach needs human interaction when there is something not working. For the approach where the rainspub must provide the delegations we need something additional, which I would prefer to not have. What is your thought, about the alarm approach?

---

*[DESIGN-MISSING-PUBLICKEY.md, line 32 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN91MropzI-aZYvHgM:-KqNqvzR6QgafN_7WDE2:b2ss7xm) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L32)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

what? no. rainspub is one-way.
</blockquote></details>

see comment above

---

*[DESIGN-MISSING-PUBLICKEY.md, line 44 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN9R7vMAgL-1PA3RxB:-KqNr-vO_yW-yCNzTNYv:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L44)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

s/go routine/goroutine/
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 50 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNtlXk2p1mPkzjQ84S:-KqNtlXk2p1mPkzjQ84T:b-6j1csb) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L50)):*
> ```Markdown
> Depending on the answer the server receives as a response to the query it handles the pending
> section differently. It first removes the token from the active token cache and sets the value of a
> map which is keyed by the token to the number of assertions containing a delegation object. It then
> ```

According to our discussion we also do not wait until we have all delegations sent in a message as an answer to a delegation query. Firstly, because they can arrive in different messages, secondly, because they could arrive independently of the query. Thirdly, it removes complexity. Right?

---

*[DESIGN-MISSING-PUBLICKEY.md, line 57 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN9vU-NL1WCrGxcQux:-KqNug3webiBtZaZO3jJ:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L57)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

yes! to avoiding unnecessary special cases.
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 70 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNAAhZ7cmisHCU9dQ9:-KqNvoTebeRuGW7x2wHg:b-sg83fd) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L70)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

remember, blacklisting is a separate process.
</blockquote></details>

Done. moved to a separate file where I collect thoughts and discussions we will have or had about blacklisting

---

*[DESIGN-MISSING-PUBLICKEY.md, line 83 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNANHKLoqpx-VUutHY:-KqNwz0OMF26Hm7IXZUw:b8s2qar) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L83)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

a bad signature on a delegation isn't necessarily an inconsistency event
</blockquote></details>

Yes that is correct. But the removal is related to the constraint we got ride of, that a server either has no delegation information or everything.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 84 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNAfzzrFwCqY0QmwFw:-KqNyT2nDngAp6BhSazS:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L84)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

...more detail here please.
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 85 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNAZDeJjHZhicD9SsH:-KqNwkkRsvcNu5G0CwbA:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L85)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

s/silent/silently.
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 87 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNAuwhgzS2xbklPDYB:-KqNxsGi6Db6fmP4vBzS:b1ojh7c) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L87)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

this logic should appear in the section in the redirect cache, I think...
</blockquote></details>

The logic belongs here because it handles a missing public key. But similar to the design-handle-queries.md it should decide upon a global config if it should follow a redirect or send it back and use the redirect cache to obtain appropriate information

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/44)*

britram:
Reviewed 7 of 8 files at r2.
Review status: 12 of 92 files reviewed at latest revision, 13 unresolved discussions.

---

*[DESIGN-BLACKLIST-HEURISTICS.md, line 9 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX6TXTYkfOPhcnw_W1:-KqX6TXUnSeyMKB9QmjU:b-gw4mdt) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-BLACKLIST-HEURISTICS.md#L9)):*
> ```Markdown
> - Assertion(s) in response to a delegation query not all containing either a public key, an IP4, an
>   IP6 or a redirect object for the queried zone are extracted from the logs and the zone's possibly
>   blacklisted. We do not want to allow normal assertion(s) to take the path over the priority queue
> ```

I wouldn't overweight redirects here. Redirects are *necessary* when a RAINS infrastructure uses iterative query service and hierarchical caching as DNS, but there are other topologies where redirects aren't imperative. The blacklist policy should match the topology.

---

*[DESIGN-HANDLE-QUERIES.md, line 19 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNBB1xs98cr7qO3fYM:-KqX7NHj702XI0Odcixp:b-dq72w8) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L19)):*
<details><summary><i>Previously, fehlmach wrote…</i></summary><blockquote>

This note does not concern the cache but the handling of a delegation query in general. We decided to log in the active token cache whenever an entry expires. Based on that an external service can then decide if it wants to blacklist this server. And I note here that if the server decides to do the lookup by himself it might miss this deadline and might get blacklisted by the external service.
</blockquote></details>

OK -- although note that any blacklisting policy that blacklists on a normal deadline miss is a bad policy. I guess we can let people decide whether they want to make their own RAINS servers useless out of paranoia though. :)

---

*[DESIGN-HANDLE-QUERIES.md, line 40 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNEQGXPNc9EIKJoUKL:-KqX7hYwY2G5goRoYV1x:b-zej12n) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L40)):*
<details><summary><i>Previously, fehlmach wrote…</i></summary><blockquote>

We could mark delegation assertions as internal in the assertion cache, then they get only evicted when they expire or the zone has sent an inconsistent message (in which case all sections from this zone are removed) What do you think about that? then we can handle them the same as any other query.
</blockquote></details>

LGTM yep.

---

*[DESIGN-HANDLE-QUERIES.md, line 6 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX72_oy4Es2nBrfPhM:-KqX72_oy4Es2nBrfPhN:b-aj79s8) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-HANDLE-QUERIES.md#L6)):*
> ```Markdown
> - Delegation queries are handled slightly differently because they are important for the system to
>   run with minimal disruption.
> - Note that if a rains server decides to do the lookup by himself for a delegation query it might be
> ```

s/by himself/itself/ (nonsentient entities are genderless in English, with a few, usually nautical, exceptions)

---

*[DESIGN-MISSING-PUBLICKEY.md, line 9 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN3-9Vp8DGhbRZOXrq:-KqX7sKzmgn-zk7Zfoqp:b-i2u9t9) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L9)):*
<details><summary><i>Previously, fehlmach wrote…</i></summary><blockquote>

The main reason was that I wanted to wait until all delegation assertions came in before we answer the query. But as we now answer as soon as we have any information, we can remove this part. Concerning complexity, we can design the pending signature cache such that it takes one hashmap lookup + cleanup of activeToken cache.
</blockquote></details>

See my comment on damping in the general design decisions...

---

*[DESIGN-MISSING-PUBLICKEY.md, line 24 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN7eZlDJL_Fqhympq6:-KqX82c9uxe4B6LFumGB:blts4hb) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L24)):*
<details><summary><i>Previously, fehlmach wrote…</i></summary><blockquote>

What is the rains doc?
</blockquote></details>

draft-trammell-rains-protocol -- this was a note to self (now reflected in an issue on that repo)

---

*[DESIGN-MISSING-PUBLICKEY.md, line 27 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN8ZHyto3_gbkd7fTL:-KqX8F-tRYhsgUMafjCo:b-ezqzp5) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L27)):*
<details><summary><i>Previously, fehlmach wrote…</i></summary><blockquote>

I also though about this approach. The reason why I did not choose it was that rainspub must get this information beforehand. out-of-band or by a client run at the authority to query the super ordinate zone for the delegation.  In the case I described everything works in-band. Instead of sending something back to the rainspub we could just log an error message that the received content from rainspub could not be processed. We should IMO also raise an alarm such that an operator can check what the problem is before the zone gets unavailable. The alarm approach needs human interaction when there is something not working. For the approach where the rainspub must provide the delegations we need something additional, which I would prefer to not have. What is your thought, about the alarm approach?
</blockquote></details>

let's resolve at the whiteboard

---

*[DESIGN-MISSING-PUBLICKEY.md, line 50 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNtlXk2p1mPkzjQ84S:-KqX8NiLNjtXZhrL_lD7:brxg05f) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L50)):*
<details><summary><i>Previously, fehlmach wrote…</i></summary><blockquote>

According to our discussion we also do not wait until we have all delegations sent in a message as an answer to a delegation query. Firstly, because they can arrive in different messages, secondly, because they could arrive independently of the query. Thirdly, it removes complexity. Right?
</blockquote></details>

yes, however, see other comment on damping.

---

*[DESIGN-NOTES.md, line 3 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX51iRxyMRTTxQRY3T:-KqX51iSfh5Mr31dQ8jK:b-riwmd5) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-NOTES.md#L3)):*
> ```Markdown
> # General Design decisions
> - Queries are answered as soon as any queried information is available, i.e. a server does not wait
>   until it can fully answer a query but already responds with partial information.
> ```

Actually, we probably want a *little* delay here -- two subordinate queries that return within 10ms of each other should probably result in a single answer, within 2000ms not. From a control standpoint, there needs to be some damping of the system.

---

*[DESIGN-NOTES.md, line 5 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX5EDIwMKgnbZPWBxW:-KqX5EDIwMKgnbZPWBxX:b-e6p3af) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-NOTES.md#L5)):*
> ```Markdown
>   until it can fully answer a query but already responds with partial information.
> - A server should keep all public keys necessary to validate the signatures on all its cached
>   sections. This allows the server to answer all delegation queries about sections it has sent out.
> ```

Yes; notably, it should keep all the sections around it needs to prove the chain.

---

*[DESIGN-NOTES.md, line 10 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX5wGkt2oEBQdmSaVR:-KqX5wGkt2oEBQdmSaVS:bdcni05) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-NOTES.md#L10)):*
> ```Markdown
>   already sent to another server.
> - After a configurable maximum number of delegation requests not leading to a needed public key for
>   a section, it will be dropped.
> ```

this is not clear. "delegation request" -> "a query sent from server A to server B", "it will be dropped"...meaning server A will stop asking?

---

*[DESIGN-NOTES.md, line 12 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX6CBwfG4aeSWdW2cH:-KqX6CBwfG4aeSWdW2cI:b-b15j4i) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-NOTES.md#L12)):*
> ```Markdown
>   a section, it will be dropped.
> - A server should log all incoming messages and their content such that an external service can
>   blacklist zones or IP addresses based on it and additional information.
> ```

And there should be a method for updating blacklist peers / zones at runtime; I'm envisioning a coprocess which reacts to rainsd logging in real time for DoS defense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/44)*

fehlmach:
Review status: 12 of 92 files reviewed at latest revision, 11 unresolved discussions.

---

*[DESIGN-BLACKLIST-HEURISTICS.md, line 9 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX6TXTYkfOPhcnw_W1:-KqXDHLyNGpF5lTyj0m3:b-er9hui) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-BLACKLIST-HEURISTICS.md#L9)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

I wouldn't overweight redirects here. Redirects are *necessary* when a RAINS infrastructure uses iterative query service and hierarchical caching as DNS, but there are other topologies where redirects aren't imperative. The blacklist policy should match the topology.
</blockquote></details>

How are redirects overweighted here? This point just wants to make sure that one cannot abuse the priority queue by sending different content than answers to delegations.

---

*[DESIGN-HANDLE-QUERIES.md, line 19 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNBB1xs98cr7qO3fYM:-KqXEV84CElBv2XEtCqm:b9wzflv) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L19)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

OK -- although note that any blacklisting policy that blacklists on a normal deadline miss is a bad policy. I guess we can let people decide whether they want to make their own RAINS servers useless out of paranoia though. :)
</blockquote></details>

yes of course. But firstly, only deadline misses on delegation queries are taken into consideration which are crucial for the system to work and secondly, it should only blacklist a server if this behavior happens frequently not just once. Also depending on the server topology a server might be white listed because he is expected to do an iterative lookup.

---

*[DESIGN-HANDLE-QUERIES.md, line 40 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNEQGXPNc9EIKJoUKL:-KqXGXAx43tiDoyFY5Y-:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-HANDLE-QUERIES.md#L40)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

LGTM yep.
</blockquote></details>

Done.

---

*[DESIGN-HANDLE-QUERIES.md, line 6 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX72_oy4Es2nBrfPhM:-KqXE9hh0_uvrb0nA2ON:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-HANDLE-QUERIES.md#L6)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

s/by himself/itself/ (nonsentient entities are genderless in English, with a few, usually nautical, exceptions)
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 9 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN3-9Vp8DGhbRZOXrq:-KqXGfQAuClrPcqK-Osm:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L9)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

See my comment on damping in the general design decisions...
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 24 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN7eZlDJL_Fqhympq6:-KqXHBaJ045nrigJis8A:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L24)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

draft-trammell-rains-protocol -- this was a note to self (now reflected in an issue on that repo)
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 27 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqN8ZHyto3_gbkd7fTL:-KqXHxh49b0DBtD3KeyW:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L27)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

let's resolve at the whiteboard
</blockquote></details>

Done.

---

*[DESIGN-MISSING-PUBLICKEY.md, line 50 at r1](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqNtlXk2p1mPkzjQ84S:-KqXI2j1OefmbNDE16Bt:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/22eb3172ce6996a5128c98067f7d6fe25ebf95af/DESIGN-MISSING-PUBLICKEY.md#L50)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

yes, however, see other comment on damping.
</blockquote></details>

Done.

---

*[DESIGN-NOTES.md, line 3 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX51iRxyMRTTxQRY3T:-KqXNw_1nBbEBFUpdJrt:b-exjpns) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-NOTES.md#L3)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

Actually, we probably want a *little* delay here -- two subordinate queries that return within 10ms of each other should probably result in a single answer, within 2000ms not. From a control standpoint, there needs to be some damping of the system.
</blockquote></details>

I'll write that up.

---

*[DESIGN-NOTES.md, line 5 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX5EDIwMKgnbZPWBxW:-KqXIICIJfmpiEM6-AOU:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-NOTES.md#L5)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

Yes; notably, it should keep all the sections around it needs to prove the chain.
</blockquote></details>

Done.

---

*[DESIGN-NOTES.md, line 10 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX5wGkt2oEBQdmSaVR:-KqXIcGjuIUOeFamNMtt:bv3j3k9) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-NOTES.md#L10)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

this is not clear. "delegation request" -> "a query sent from server A to server B", "it will be dropped"...meaning server A will stop asking?
</blockquote></details>

I update it according to our discussion

---

*[DESIGN-NOTES.md, line 12 at r2](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqX6CBwfG4aeSWdW2cH:-KqXNp2aY26A-6-zpEey:b-7tfmjp) ([raw file](https://github.com/netsec-ethz/rains/blob/c2dcc43ce07ef06f6ebe2daf3517c7e53df2f39d/DESIGN-NOTES.md#L12)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

And there should be a method for updating blacklist peers / zones at runtime; I'm envisioning a coprocess which reacts to rainsd logging in real time for DoS defense.
</blockquote></details>

added

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/44)*

britram:
Review status: 4 of 91 files reviewed at latest revision, 13 unresolved discussions.

---

*[DESIGN-NOTES.md, line 18 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcAOQbfuTyvd285lxD:-KqcAOQbfuTyvd285lxE:bvne27u) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L18)):*
> ```Markdown
>   queries as well as in the zone key cache which is internally used to lookup public keys.
> - It must be possible to update peer and zone blacklists at runtime such that an external service or
>   coprocess analyzing rainsd's logging in real time can defend DoS attacks.
> ```

OK, note not just DoS

---

*[DESIGN-NOTES.md, line 30 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcATr4WSL4FRJEo93Q:-KqcATr4WSL4FRJEo93R:b-kzp3qb) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L30)):*
> ```Markdown
> 
> ## Section and Message processing proposal
> - The number of concurrently active goroutines working on messages are restricted to a configurable
> ```

this presumes a one-goroutine-per-message (or rather, per section) design, yes?

---

*[DESIGN-NOTES.md, line 38 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcAeXLMqio2M1oLctZ:-KqcAeXLMqio2M1oLct_:b-ocw0h1) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L38)):*
> ```Markdown
> - Goroutines working on sections maintain two channels to the goroutine of the message, one to
>   signal back when it is done and the other to signal back when it has to wait for an
>   answer.
> ```

This seems to be somewhat more complicated than it needs to be.

Counterproposal: when a goroutine hits a waitpoint, it places the thing it's waiting on in the appropriate cache and exits. When the thing it's waiting on arrives, the routine handling the arrival removes it from the appropriate cache and spawns a new goroutine, as if the pending section has just arrived.

---

*[DESIGN-NOTES.md, line 41 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcB5tQAqPjbfSSyiNe:-KqcB5tQAqPjbfSSyiNf:bge1115) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L41)):*
> ```Markdown
> - In case a section goroutine has to wait for a query's response, it adds a channel to the
>   appropriate pending cache and blocks on it. When an answer arrived the pending cache signals the
>   goroutine that it can continue its processing.
> ```

Again, why can't we let things rest in the pending caches?

---

*[DESIGN-NOTES.md, line 46 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcBSzO0DALAdVJd4ys:-KqcBSzO0DALAdVJd4yt:bchkrmh) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L46)):*
> ```Markdown
>     wait-time has passed (e.g. 10ms). Each time a message with the same token arrives the countdown
>     is reset to the configured wait-time. In case the incoming message is not a response to a query
>     it goes straight to step 3.
> ```

does the engine do this?

---

*[DESIGN-NOTES.md, line 50 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcBbRyBllkvc60KW0N:-KqcBbRyBllkvc60KW0O:bak2fxs) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L50)):*
> ```Markdown
>   3. The message goroutine creates for each contained section(s) a new goroutine
>   4. Depending on the configuration one of the following processing steps is performed.
>     - The message goroutine waits until all sections are successfully processed. It then checks if
> ```

OK,  there's a message goroutine which spawns and waits on section goroutines... that makes snense.

---

*[DESIGN-NOTES.md, line 55 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcBkJC1olTiKQgE_4T:-KqcBkJDHW-43eFrC6-3:b6khitl) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L55)):*
> ```Markdown
>     - The message goroutine directly forwards the message to all pending queries or waits until the
>       first section is done to invoke the goroutines waiting for a public key. Then it waits for all
>       sections to terminate and decreases the message goroutine counter.
> ```

What is the message goroutine doing other than eating state?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/44)*

fehlmach:
Review status: 4 of 91 files reviewed at latest revision, 13 unresolved discussions.

---

*[DESIGN-NOTES.md, line 18 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcAOQbfuTyvd285lxD:-Kqg-HLe2AQJ0Id8U6Vw:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L18)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

OK, note not just DoS
</blockquote></details>

Done.

---

*[DESIGN-NOTES.md, line 30 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcATr4WSL4FRJEo93Q:-Kqg-KdyX_AUGfIxFx-V:b2nign6) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L30)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

this presumes a one-goroutine-per-message (or rather, per section) design, yes?
</blockquote></details>

yes

---

*[DESIGN-NOTES.md, line 38 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcAeXLMqio2M1oLctZ:-Kqg1O-PFk5yDiUw2FyY:bqhs7je) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L38)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

This seems to be somewhat more complicated than it needs to be.

Counterproposal: when a goroutine hits a waitpoint, it places the thing it's waiting on in the appropriate cache and exits. When the thing it's waiting on arrives, the routine handling the arrival removes it from the appropriate cache and spawns a new goroutine, as if the pending section has just arrived.
</blockquote></details>

This would work if all sections are completely independent of each other. But we want to be able to answer some queries with several sections together. So there must be some common thing between them, a channel.  I try to structure it a bit better so that it is more comprehensible. If you are still not convinced, we can discuss the disadvantages and advantages of both.

---

*[DESIGN-NOTES.md, line 41 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcB5tQAqPjbfSSyiNe:-Kqg-uFme2cMu-iXzSNK:bmetdo3) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L41)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

Again, why can't we let things rest in the pending caches?
</blockquote></details>

because the message goroutine is waiting until all section goroutines are finished. But this one has to wait. I'd suggest to make it configurable when the section goroutine signals the message goroutine that it is done either when it has to wait or when it got the answer.  With this approach an operator can decide if the message goroutine waits until all information is ready or already answers query after partial information is available.

---

*[DESIGN-NOTES.md, line 46 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcBSzO0DALAdVJd4ys:-Kqg-OrmO5UAwgrbWzss:bi2c78g) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L46)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

does the engine do this?
</blockquote></details>

yes, but in the design I should only talk about the server. 

---

*[DESIGN-NOTES.md, line 50 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcBbRyBllkvc60KW0N:-Kqg-pM7ovvidAdeAKhJ:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L50)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

OK,  there's a message goroutine which spawns and waits on section goroutines... that makes snense.
</blockquote></details>

Done.

---

*[DESIGN-NOTES.md, line 55 at r3](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KqcBkJC1olTiKQgE_4T:-Kqg-jJMmB6hpCKfgojf:baqno5w) ([raw file](https://github.com/netsec-ethz/rains/blob/cf643dd19b0e41a0891fba90c08516492d76162a/DESIGN-NOTES.md#L55)):*
<details><summary><i>Previously, britram (Brian Trammell) wrote…</i></summary><blockquote>

What is the message goroutine doing other than eating state?
</blockquote></details>

handling the sections and queries in the pending caches

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/44)*

britram:
Reviewed 1 of 9 files at r4.
Review status: 5 of 91 files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/44)*

britram:
Reviewed 5 of 19 files at r3, 3 of 25 files at r5.
Review status: 10 of 96 files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/44)*

britram:
:lgtm:

---

Reviewed 59 of 105 files at r1, 4 of 19 files at r3, 2 of 9 files at r4, 6 of 25 files at r5, 44 of 45 files at r6.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[rainslib/datamodel.go, line 109 at r6](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-KtGTqLgQmtfQg_4_FWQ:-KtGTqLgQmtfQg_4_FWR:b-7fr1w7) ([raw file](https://github.com/netsec-ethz/rains/blob/9d0160d2e8311be0e970211a0928883f15a38d43/rainslib/datamodel.go#L109)):*
> ```Go
> const (
> 	//NoCapability is used when the server does not listen for any connections
> 	NoCapability Capability = "urn:x-rains:nocapability"
> ```

FYI this isn't as specified; will file an issue.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/netsec-ethz/rains/44#-:-KtGVcWmP3O39x4tTmhp:bnfp4nl)*

Issue #117, opened 2018-12-11
Move cache implementations to own package


fehlmach:
Currently part of the rainsd package

britram:
closed by #143

PR #143, opened 2019-01-07
Separate package for caches

fehlmach:
Moves cache implementation from the rainsd package to an own 'cache' package. This way it can be used in the libresolve as well.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/143)


britram:
fixes #117

PR #145, opened 2019-01-07
Replaces custom network address with go's stdlib net.Addr

fehlmach:
The custom network address object 'ConnInfo' is still there to simplify json parsing


britram:
will fix #116 on merge

PR #157, opened 2019-01-07
Rainsd SCION listener support.

rayhaanj:
This pull requests implements functionality to listen on SCION addresses in rainsd.

It does not use the connection cache because the scion transport is using UDP and hence has no notion of connections. For this reason we just put the `net.Conn` in the address to send back to, and that is used in `sendTo`.

It also adds support for reading SCION addresses from the configuration file.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/157)


britram:
This is a pretty deep change, more or less requiring the whole SCION codebase to be pulled in to build and test whether SCION will be configured or not. For SCIONlab, it's fine, but it would be nice if this could be conditionally built.

Looking at the codebase, and how the connection stuff is designed, I'm not sure how easy that would be.

@fehlmach are there other things that need to happen in your codebase before this can land?

fehlmach:
There should be integration tests which check the scion connectivity between servers.
There probably are some switch statements which would hit the default case (error case) which must be extended. The best way to find them is probably to have an integration test and fix one after the other until it runs.
The recursive lookup in the resolver also needs a little change. There are already TODO's placed.

PR #170, opened 2019-01-16
119 check signatures in rlookup

### LINKED ISSUE #119: Check signatures of answers in a recursive lookup in resolver
Marked with FIXME in resolver.go
1) check if delegation was sent together with answer.
1a) if not send a new query for the delegation back to the sender. 
2) check signature using siglib
2a) stop processing in case signature is invalid.

juagargi:
Closes #119 

Needs UT and integration tests.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/170)

### DIFF
--- a/cmd/rainsd/rainsd.go
+++ b/cmd/rainsd/rainsd.go
@@ -195,7 +195,6 @@ func main() {
 		log.Fatal(err)
 	}
 	if !rootCmd.Flag("help").Changed {
-
 		updateConfig(&config)
 		server, err := rainsd.New(config, id)
 		if err != nil {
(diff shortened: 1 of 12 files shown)


fehlmach:
-> Contained section has a subjectZone or context
For signing we add the context and zone to assertions contained in a zone or a shard. They should be removed afterwards but somewhere that is not the case. I guess you do not do it in the resolver. 
The reason CheckSignature does not do it, is because we want to cache the assertions and then this information must be present. So in the server we remove it after the assertions are cached.

PR #175, opened 2019-01-17
Add SCION address object types.

rayhaanj:
This commit adds the object types OTScionAddr4, OTScionAddr6 with
type 14, 15 respectively.

Also adds the appropriate encoder and decoder methods for CBOR. The
CBOR representation uses the snet utilities for converting addresses
to and from string representation.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/175)


britram:
This seems like it's off the same branch as #157, so master should merge back to this after it lands (and I'll review it then)

PR #212, opened 2019-01-31
Added two UTs to libresolve

juagargi:
# Do not merge yet

Review the approach

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/212)

### DIFF
--- a/internal/pkg/libresolve/resolver.go
+++ b/internal/pkg/libresolve/resolver.go
@@ -38,6 +38,15 @@ const (
 	Forward
 )
 
+// some of these types are not "method expressions" but will be invoked as such
+// they (or an interface-based approach) are needed to decouple logic and run tests on different
+// parts of the Resolver type
+
+type sendQueryFcnType func(msg message.Message, addr net.Addr, timeout time.Duration) (message.Message, error)
+type handleAnswerFcnType func(r *Resolver, msg message.Message, q *query.Name, recurseCount int) (
+	isFinal bool, isRedir bool, redirMap map[string]string, srvMap map[string]object.ServiceInfo,
+	ipMap map[string]string, nameMap map[string]object.Name)
+
 // Resolver provides methods to resolve names in RAINS.
 type Resolver struct {
 	RootNameServers   []net.Addr
@@ -50,6 +59,8 @@ type Resolver struct {
 	Connections       cache.Connection
 	MaxCacheValidity  util.MaxCacheValidity
 	MaxRecursiveCount int
+	sendQueryFcn      sendQueryFcnType
+	handleAnswerFcn   handleAnswerFcnType
 }
 
 //New creates a resolver with the given parameters and default settings
@@ -65,6 +76,9 @@ func New(rootNS, forwarders []net.Addr, rootKeyPath string, mode ResolutionMode,
 		Connections:       cache.NewConnection(maxConn),
 		MaxCacheValidity:  maxCacheValidity,
 		MaxRecursiveCount: maxRecursiveCount,
+		// now the pointers to functions
+		sendQueryFcn:    util.SendQuery,
+		handleAnswerFcn: _handleAnswer,
 	}
 	// load the root zone public key and store it as a delegation:
 	a := new(section.Assertion)
@@ -149,7 +163,7 @@ func (r *Resolver) forwardQuery(q *query.Name) (*message.Message, error) {
 	}
 	for _, forwarder := range r.Forwarders {
 		msg := message.Message{Token: token.New(), Content: []section.Section{q}}
-		answer, err := util.SendQuery(msg, forwarder, r.DialTimeout*time.Millisecond)
+		answer, err := r.sendQueryFcn(msg, forwarder, r.DialTimeout*time.Millisecond)
 		if err == nil {
 			return &answer, nil
 		}
@@ -179,7 +193,7 @@ func (r *Resolver) recursiveResolve(q *query.Name, recurseCount int) (*message.M
 		addr := root
 		for {
 			msg := message.Message{Token: token.New(), Content: []section.Section{q}}
-			answer, err := util.SendQuery(msg, addr, r.DialTimeout*time.Millisecond)
+			answer, err := r.sendQueryFcn(msg, addr, r.DialTimeout*time.Millisecond)
 			if err != nil || len(answer.Content) == 0 {
 				break
 			}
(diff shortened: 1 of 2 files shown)


internal/pkg/libresolve/resolver.go
```diff
@@ -38,6 +38,15 @@ const (
 	Forward
 )
 
+// some of these types are not "method expressions" but will be invoked as such
+// they (or an interface-based approach) are needed to decouple logic and run tests on different
+// parts of the Resolver type
+
+type sendQueryFcnType func(msg message.Message, addr net.Addr, timeout time.Duration) (message.Message, error)
```
rayhaanj: I think that a name such as `QuerySender` is clearer for this type, and `AnswerHandler` for the one below

internal/pkg/libresolve/resolver.go
```diff
@@ -149,7 +163,7 @@ func (r *Resolver) forwardQuery(q *query.Name) (*message.Message, error) {
 	}
 	for _, forwarder := range r.Forwarders {
 		msg := message.Message{Token: token.New(), Content: []section.Section{q}}
-		answer, err := util.SendQuery(msg, forwarder, r.DialTimeout*time.Millisecond)
+		answer, err := r.sendQueryFcn(msg, forwarder, r.DialTimeout*time.Millisecond)
```
rayhaanj: It would still make sense if it was called `sendQuery` because it's clear we're calling it

internal/pkg/libresolve/resolver.go
```diff
@@ -208,11 +222,16 @@ func (r *Resolver) recursiveResolve(q *query.Name, recurseCount int) (*message.M
...
 func (r *Resolver) handleAnswer(msg message.Message, q *query.Name, recurseCount int) (isFinal bool, isRedir bool,
+	redirMap map[string]string, srvMap map[string]object.ServiceInfo, ipMap map[string]string, nameMap map[string]object.Name) {
+	return r.handleAnswerFcn(r, msg, q, recurseCount)
+}
+
+// _handleAnswer stores delegation assertions in the delegationCache. It informs the caller if msg
+// answers q. It also returns if the msg contains a redirect assertion which indicates that
+// another lookup must be performed. Information that is relevant for the next lookup are returned in
+// maps.
+func _handleAnswer(r *Resolver, msg message.Message, q *query.Name, recurseCount int) (isFinal bool, isRedir bool,
```
rayhaanj: It's fine if this is just called handleAnswer without the underscore, that would fit in with the rest of the codebase.

internal/pkg/libresolve/resolver_test.go
```diff
@@ -1,9 +1,15 @@
 package libresolve
 
 import (
+	"log"
```
rayhaanj: The rest of the codebase is using `log15` so this should be changed to `log "github.com/inconshreveable/log15"`

internal/pkg/libresolve/resolver_test.go
```diff
@@ -31,7 +37,10 @@ func TestRecursiveResolve(t *testing.T) {
 		},
 		MaxRecursiveCount: 1,
 	}
-	q := query.Name{
+}
+
+func createQuery() *query.Name {
```
rayhaanj: For consistency `newQuery`

internal/pkg/libresolve/resolver_test.go
```diff
@@ -13,8 +19,8 @@ import (
 	"github.com/netsec-ethz/rains/internal/pkg/util"
 )
 
-func TestRecursiveResolve(t *testing.T) {
-	resolver := &Resolver{
+func createResolver() *Resolver {
```
rayhaanj: For consistency `newResolver`

internal/pkg/libresolve/resolver_test.go
```diff
@@ -40,14 +49,50 @@ func TestRecursiveResolve(t *testing.T) {
...
+	_, err = resolver.recursiveResolve(q, 0)
 	if err == nil || strings.HasPrefix(err.Error(), "Maximum number of recursive calls") {
 		t.Errorf("Unexpected error about max. recursive calls. This is the error: %v", err)
 	}
 }
+
+func TestRecursiveResolveQueriesRoot(t *testing.T) {
+	assertion := section.Assertion{SubjectZone: ".", SubjectName: "ch"}
+	resolver := createResolver()
+	resolver.RootNameServers = []net.Addr{&net.IPAddr{IP: net.IPv4(127, 0, 0, 1), Zone: "test-zone"}}
```
rayhaanj: Zones are used for link local disambiguation, is there a specific reason that this is `test-zone` here?

internal/pkg/libresolve/resolver_test.go
```diff
@@ -40,14 +49,50 @@ func TestRecursiveResolve(t *testing.T) {
...
+	numberOfMessagesSent := 0
+	resolver.sendQueryFcn = func(msg message.Message, addr net.Addr, timeout time.Duration) (message.Message, error) {
+		if addr.(*net.IPAddr) == nil || addr.(*net.IPAddr).Zone != "test-zone" {
```
rayhaanj: I don't think this works to check if the IPAddr is nil: https://play.golang.org/p/6iFrxBK02wE

If you want to check if addr is an IPAddr or not then:
```go
if addr, ok := addr.(*net.IPAddr); ok {
// ...
}
```

But if you only care if it is nil, then just compare to nil, no cast

internal/pkg/libresolve/resolver_test.go
```diff
@@ -40,14 +49,50 @@ func TestRecursiveResolve(t *testing.T) {
...
+	resolver.handleAnswerFcn = func(r *Resolver, msg message.Message, q *query.Name, recurseCount int) (
+		isFinal bool, isRedir bool, redirMap map[string]string, srvMap map[string]object.ServiceInfo,
+		ipMap map[string]string, nameMap map[string]object.Name) {
+		isFinal = true
+		return
+	}
+	q := createQuery()
+	ans, err := resolver.recursiveResolve(q, 0)
+	if err != nil {
+		t.Fatalf("Unexpected error: %v", err)
```
rayhaanj: Clearer to say what failed:

`recursiveResolve failed: %v`

internal/pkg/libresolve/resolver_test.go
```diff
@@ -40,14 +49,50 @@ func TestRecursiveResolve(t *testing.T) {
...
+	}
+	if len(ans.Content) != 1 || ans.Content[0].(*section.Assertion) == nil || ans.Content[0].(*section.Assertion).FQDN() != assertion.FQDN() {
+		t.Errorf("Wrong answer received, fqdn=%s", assertion.FQDN())
```
rayhaanj: You can use %q here to make it consistent with colon notation instead of equals: `FQDN: %q` and then it will appear in quotes in the output.

PR #215, opened 2019-02-14
Enable running rains over SCION connections

FR4NK-W:
Enable running rains over SCION connections
Do not cache SCION connections

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/215)


internal/pkg/connection/connection.go
```diff
@@ -95,33 +103,89 @@ func CreateConnection(addr net.Addr) (conn net.Conn, err error) {
...
 	pathMgr := snet.DefNetwork.PathResolver()
-	pathSet := pathMgr.Query(ctx, la.IA, ra.IA)
+	pathSet := pathMgr.Query(ctx, local.IA, remote.IA)
 	for _, p := range pathSet {
 		return p.Entry
 	}
 	return nil
 }
 
+func getLocalIP() (net.IP, error) {
```
rayhaanj: This seems a bit hardcoded to work with the current setup. I guess it's fine for now but if the adresses change then this would need to be updated. I would try to make the VPN endpoint a variable/flag with this as the default value.

More ideally (but maybe not possible right now), there would be an interface called `scion0` or something and then you can use a library like https://godoc.org/github.com/vishvananda/netlink to find that interface and get the local address out.

What is the local IP used for?

internal/pkg/libresolve/resolver.go
```diff
@@ -39,6 +40,8 @@ const (
 	Forward
 )
 
+const maxUDPPacketBytes = 9000
```
rayhaanj: this is already defined in internal/pkg/connection/connection.go, but is private. It would be good to make it public and put it in one place for easier updating.

internal/pkg/libresolve/resolver.go
```diff
@@ -377,22 +393,39 @@ func (r *Resolver) handleZone(z *section.Zone, redirMap map[string]string,
...
-			return net.ResolveTCPAddr("", fmt.Sprintf("%s:%d", ipAddr, rainsPort))
-		}
-	}
-	if allowedTypes[object.OTScionAddr6] || allowedTypes[object.OTScionAddr4] {
-		if ipAddr, ok := ipMap[name]; ok {
-			return snet.AddrFromString(fmt.Sprintf("%s:%d", ipAddr, rainsPort))
+			var addr net.Addr
+			addr, err := net.ResolveTCPAddr("", fmt.Sprintf("%s:%d", ipAddr, rainsPort))
+			if err != nil {
+				log.Info("ResolveTCPAddr: Not an IP addr, will try as an SCION address", "addr", addr, "ipAddr", ipAddr, "err", err)
```
rayhaanj: Log spam can be reduced by trying both then only printing an error once with why TCP failed and why SCION failed.

internal/pkg/libresolve/resolver.go
```diff
@@ -377,22 +393,39 @@ func (r *Resolver) handleZone(z *section.Zone, redirMap map[string]string,
...
+			var err error
+			if addr, err = r.handleRedirect(srvVal.Name, srvMap, ipMap, nameMap,
 				AllowedAddrTypes); err == nil {
-				ip := strings.Split(addr.String(), ":")[0]
-				return net.ResolveTCPAddr("", fmt.Sprintf("%s:%d", ip, srvVal.Port))
+				portSep := strings.LastIndex(addr.String(), ":")
+				ip := addr.String()[:portSep]
+				addr, err = net.ResolveTCPAddr("", fmt.Sprintf("%s:%d", ip, srvVal.Port))
+				if err != nil {
+					log.Info("ResolveTCPAddr: Not an IP addr, will try as an SCION address", "addr", addr, "err", err)
```
rayhaanj: Same as above

internal/pkg/util/rainslibUtil.go
```diff
@@ -137,9 +139,22 @@ func SendQuery(msg message.Message, addr net.Addr, timeout time.Duration) (
...
-	if err := writer.Marshal(&msg); err != nil {
-		return message.Message{}, fmt.Errorf("failed to marshal message: %v", err)
+	switch addr.(type) {
+	case *net.TCPAddr:
+		writer := cbor.NewWriter(conn)
+		if err := writer.Marshal(&msg); err != nil {
+			return message.Message{}, fmt.Errorf("failed to marshal message: %v", err)
+		}
+	case *snet.Addr:
+		encoding := new(bytes.Buffer)
```
rayhaanj: Is this needed to put all of the bytes into one UDP packet?

test/integration/fullCoverageSCION_test.go
```diff
@@ -0,0 +1,328 @@
...
+
+	time.Sleep(500 * time.Millisecond)
+	log.Info("begin sending queries which should be cached by recursive lookup")
+	for i, query := range queries {
+		sendQueryVerifyResponse(t, *query, cachingResolver.Addr(), answers[i])
+	}
+	log.Warn("Done sending queries for cached entries from a recursive lookup")
+
+	//Restart caching resolver from checkpoint
+	time.Sleep(1000 * time.Millisecond) //make sure that caches are checkpointed
```
rayhaanj: Is there another way to check if it is ready?

PR #217, opened 2019-03-06
Make internal packages public

chaehni:
In order for external applications to make use of RAINS, library functions cannot live inside `internal` paths. 
This PR proposes to move the content of `/internal/pkg/` to `/pkg` such that it can be used externally.

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/217)


chaehni:
cloded in favor of #216

PR #240, opened 2022-02-23
Support latest SCIONLab update.

juagargi:
Changes in addr.IA (it is now a uint64).
The spath library is removed.
A snet.UDPAddr cannot be used to write without a non-nil path.
Added SCION connectivity test.
TODO:
- [x] Merge PR https://github.com/netsec-ethz/scion/pull/110
- [x] Check in `go.mod` that the referenced commits are valid

---
This change is [Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/240)


matzf (commented):
> Reviewed 7 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/240)* status: 7 of 9 files reviewed, 2 unresolved discussions / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[internal/pkg/connection/scion/appnet.go, line 107 at r1](https://reviewable.io/reviews/netsec-ethz/rains/240#-MwzikJk7FBoEfUlXwCx:-MwzikJk7FBoEfUlXwCy:b-by4n06) ([raw file](https://github.com/netsec-ethz/rains/blob/8b4d90ffb5fa66f963fb9758ca9a9999571f24be/internal/pkg/connection/scion/appnet.go#L107)):*
> ```Go
> // that updates the paths in case they expire or are revoked.
> func DialAddr(raddr *snet.UDPAddr) (*snet.Conn, error) {
> 	if _, ok := raddr.Path.(path.Empty); ok || raddr.Path == nil {
> ```

Nit, I think just `raddr.Path == nil` is good enough.  If the path is explicitly set to Empty, it probably does not need to be reset.
___
*[internal/pkg/connection/scion/path_selection.go, line 47 at r1](https://reviewable.io/reviews/netsec-ethz/rains/240#-Mwzj6uH4zE-fJLkp6WL:-Mwzj6uH4zE-fJLkp6WM:bpxwawt) ([raw file](https://github.com/netsec-ethz/rains/blob/8b4d90ffb5fa66f963fb9758ca9a9999571f24be/internal/pkg/connection/scion/path_selection.go#L47)):*
> ```Go
> 		p = paths[0]
> 	}
> 	SetPath(addr, p)
> ```

I find this a bit confusing to read now, because it's not obvious that `SetPath(.. nil)` will set a non-nil path. `SetPath` is only used here, so maybe we can just remove it and inline it here, :
```
	paths, err := QueryPaths(addr.IA)
	if err != nil {
		return err
	}
      else if len(paths) == 0 {
        addr.Path = path.Empty{}
        addr.NextHop = nil
      } else {
         addr.Path = paths[0].Dataplane()
         addr.NextHop = p.NextHop()
     }
```

juagargi (commented):
> *[Reviewable](https://reviewable.io/reviews/netsec-ethz/rains/240)* status: 5 of 9 files reviewed, 1 unresolved discussion / 0 of 1 LGTMs obtained / 0 of 1 approvals obtained

___
*[internal/pkg/connection/scion/appnet.go, line 107 at r1](https://reviewable.io/reviews/netsec-ethz/rains/240#-MwzikJk7FBoEfUlXwCx:-Mx9cZHaEuKCNkrhdMGQ:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/8b4d90ffb5fa66f963fb9758ca9a9999571f24be/internal/pkg/connection/scion/appnet.go#L107)):*
<details><summary><i>Previously, matzf (Matthias Frei) wrote…</i></summary><blockquote>

Nit, I think just `raddr.Path == nil` is good enough.  If the path is explicitly set to Empty, it probably does not need to be reset.
</blockquote></details>

Done.
___
*[internal/pkg/connection/scion/path_selection.go, line 47 at r1](https://reviewable.io/reviews/netsec-ethz/rains/240#-Mwzj6uH4zE-fJLkp6WL:-Mx9bO-pCWEv20VCL2dO:b-896fix) ([raw file](https://github.com/netsec-ethz/rains/blob/8b4d90ffb5fa66f963fb9758ca9a9999571f24be/internal/pkg/connection/scion/path_selection.go#L47)):*
<details><summary><i>Previously, matzf (Matthias Frei) wrote…</i></summary><blockquote>

I find this a bit confusing to read now, because it's not obvious that `SetPath(.. nil)` will set a non-nil path. `SetPath` is only used here, so maybe we can just remove it and inline it here, :
```
	paths, err := QueryPaths(addr.IA)
	if err != nil {
		return err
	}
      else if len(paths) == 0 {
        addr.Path = path.Empty{}
        addr.NextHop = nil
      } else {
         addr.Path = paths[0].Dataplane()
         addr.NextHop = p.NextHop()
     }
```
</blockquote></details>

Done.

PR #245, opened 2026-03-18
update scionproto reference to a201cf565, dated 6-02-2026

juagargi:


juagargi:
Merged without review.

